PR #29, opened 2026-04-29
fix(honesty): wire real Nexus validator stages for parity with paper-trade

### LINKED ISSUE #28: fix(honesty): wire real Nexus validator stages for parity with paper-trade
## Routing

Parent issue: #25
Depends on:
- #17

Unlocks:
- Nexus-side `ValidationRequestContext.execution_params` extension that lets the bts INTAKE pre-hooks collapse into the pipeline
- Nexus-side `CapitalController.close_position` primitive that lets SELL close run a real CAPITAL stage instead of relying on `ValidationAction.EXIT` bypass

---

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

- `.venv/bin/python -c "import ast, pathlib; tree = ast.parse(pathlib.Path('backtest_simulator/honesty/capital.py').read_text()); print('FAIL' if any(isinstance(n, ast.FunctionDef) and n.name == '_allow_stage' for n in ast.walk(tree)) else 'OK')"` returns `OK`.
- `grep -c '_allow_stage' backtest_simulator/honesty/capital.py` returns `0`.
- `.venv/bin/python -c "from decimal import Decimal; from backtest_simulator.honesty.capital import build_validation_pipeline; from nexus.instance_config import InstanceConfig; cfg = InstanceConfig(account_id='t', venue='binance_spot_simulated'); p, c, s = build_validation_pipeline(nexus_config=cfg, capital_pool=Decimal('1000')); print('FAIL' if any('_allow_stage' in p._validators[st].__qualname__ for st in p.stage_order) else 'OK')"` returns `OK`.
- `grep -cE 'return validate_intake_stage\(' backtest_simulator/honesty/capital.py` returns `1`.
- `grep -cE 'return validate_risk_stage\(' backtest_simulator/honesty/capital.py` returns `1`.
- `grep -cE 'return validate_price_stage\(' backtest_simulator/honesty/capital.py` returns `1`.
- `grep -cE 'return validate_health_stage\(' backtest_simulator/honesty/capital.py` returns `1`.
- `grep -cE 'return validate_platform_limits_stage\(' backtest_simulator/honesty/capital.py` returns `1`.
- `grep -cE 'pipeline bypassed' backtest_simulator/launcher/action_submitter.py` returns `1` (the SELL fast-path now logs the bypass reason on the lifecycle line; the bypass itself stays in place because the long-only strategy template does not propagate `Action.trade_id` to the SELL leg and `CapitalController` has no `close_position` primitive — both upstream debts).
- `.venv/bin/python -c "from backtest_simulator.launcher.action_submitter import _check_declared_stop, _check_atr_sanity; phrase='supplementing the real Nexus INTAKE stage'; print('OK' if (phrase in (_check_declared_stop.__doc__ or '') and phrase in (_check_atr_sanity.__doc__ or '')) else 'FAIL')"` returns `OK`.
- `.venv/bin/python -m pytest tests/honesty/test_validation_parity.py -q 2>&1 | tail -1 | grep -oE '[0-9]+ passed'` matches `11 passed`.
- `.venv/bin/python -m pytest -q 2>&1 | tail -1 | grep -cE 'passed'` returns `1` (whole suite still green).
- `.venv/bin/python -c "import tomllib, pathlib; print(tomllib.loads(pathlib.Path('pyproject.toml').read_text())['project']['version'])"` prints `2.0.6`.
- `head -1 CHANGELOG.md` prints `# v2.0.6`.

---

## Current State

>
> Accepted:
>
> Rejected:

- `git show main:backtest_simulator/honesty/capital.py | grep -c '^def _allow_stage('` returns `1`.
- `git show main:backtest_simulator/honesty/capital.py | grep -cE 'ValidationStage\.[A-Z_]+: _allow_stage'` returns `5`.
- `git show main:backtest_simulator/honesty/capital.py | grep -c 'validate_capital_stage'` returns `2` (one import + one call).
- `git show main:backtest_simulator/honesty/capital.py | grep -cE 'validate_(intake|risk|price|health|platform_limits)_stage'` returns `0`.
- `git show main:backtest_simulator/launcher/action_submitter.py | grep -c 'if action.direction == OrderSide.SELL:'` returns `1`.
- `git show main:backtest_simulator/launcher/action_submitter.py | grep -c 'CAPITAL skipped'` returns `1`.
- `.venv/bin/python -c "from decimal import Decimal; from backtest_simulator.honesty.capital import build_validation_pipeline; p, c, s = build_validation_pipeline(capital_pool=Decimal('1000')); print(sum(1 for st in p.stage_order if p._validators[st].__qualname__ == '_allow_stage'))"` prints `5`.
- `.venv/bin/python -c "import inspect; from backtest_simulator.honesty.capital import build_validation_pipeline; print(str(inspect.signature(build_validation_pipeline)))"` prints `(*, capital_pool: 'Decimal', reservation_ttl_seconds: 'int' = 86400) -> 'tuple[ValidationPipeline, CapitalController, CapitalState]'`.
- `python -c "import tomllib, pathlib; print(tomllib.loads(pathlib.Path('pyproject.toml').read_text())['project']['version'])"` prints `2.0.5`.

---

## Target State

> predicate.
>
> Accepted:

- `git show HEAD:backtest_simulator/honesty/capital.py | grep -c '^def _allow_stage('` returns `0`.
- `git show HEAD:backtest_simulator/honesty/capital.py | grep -c '_allow_stage'` returns `0`.
- `git show HEAD:backtest_simulator/honesty/capital.py | grep -cE 'validate_(intake|risk|price|capital|health|platform_limits)_stage\(' ` returns `≥6`.
- `git show HEAD:backtest_simulator/launcher/action_submitter.py | grep -c 'pipeline bypassed'` returns `1`.
- `.venv/bin/python -c "from decimal import Decimal; from backtest_simulator.honesty.capital import build_validation_pipeline; from nexus.instance_config import InstanceConfig; cfg = InstanceConfig(account_id='t', venue='binance_spot_simulated'); p, c, s = build_validation_pipeline(nexus_config=cfg, capital_pool=Decimal('1000')); print(sum(1 for st in p.stage_order if '_allow_stage' in p._validators[st].__qualname__))"` prints `0`.
- `.venv/bin/python -m pytest tests/honesty/test_validation_parity.py -q 2>&1 | tail -1 | grep -oE '[0-9]+ passed'` matches `11 passed`.
- `pytest -q 2>&1 | tail -1` matches `passed`.
- `.venv/bin/python -c "import tomllib, pathlib; print(tomllib.loads(pathlib.Path('pyproject.toml').read_text())['project']['version'])"` prints `2.0.6`.
- `head -1 CHANGELOG.md` prints `# v2.0.6`.
- `python -m backtest_simulator.cli gate lint 2>&1 | tail -1` prints `All checks passed!`.

---

## Surfaces

- `pyproject.toml`
- `CHANGELOG.md`
- `backtest_simulator/honesty/capital.py`
- `backtest_simulator/honesty/__init__.py`
- `backtest_simulator/launcher/launcher.py`
- `backtest_simulator/launcher/action_submitter.py`
- `README.md`
- `tests/honesty/test_capital_lifecycle.py`
- `tests/honesty/test_adapter_wrapper_paths.py`
- `tests/honesty/test_inverse_prescient.py`
- `tests/honesty/test_sanity_buy_hold.py`
- `tests/honesty/test_sanity_over_trading.py`
- `tests/honesty/test_sanity_random.py`
- `tests/honesty/test_sanity_zero_trade.py`
- `tests/honesty/test_sell_close_semantics.py`
- `tests/honesty/test_shuffle_bars.py`
- `tests/launcher/test_action_submitter.py`
- `tests/tools/test_lint_ci_contract.py`

Added:

- `tests/honesty/test_validation_parity.py`

Removed:

- (none)

---

## Signature Diff

### `backtest_simulator/honesty/capital.py`::`build_validation_pipeline`
Before:
```python
def build_validation_pipeline(
    *,
    capital_pool: Decimal,
    reservation_ttl_seconds: int = 86_400,
) -> tuple[ValidationPipeline, CapitalController, CapitalState]:
```
After:
```python
def build_validation_pipeline(
    *,
    nexus_config: InstanceConfig,
    capital_pool: Decimal,
    reservation_ttl_seconds: int = 86_400,
    risk_limits: RiskStageLimits | None = None,
    health_policy: HealthStagePolicy | None = None,
    platform_limits: PlatformLimitsStageLimits | None = None,
    health_snapshot_provider: Callable[[], HealthStageSnapshot] = _default_health_snapshot,
    platform_snapshot_provider: Callable[[], PlatformLimitsStageSnapshot] = _default_platform_snapshot,
    price_snapshot_provider: Callable[[], PriceCheckSnapshot | None] = _default_price_snapshot,
) -> tuple[ValidationPipeline, CapitalController, CapitalState]:
```

### `backtest_simulator/honesty/capital.py`::`_allow_stage`
Before:
```python
def _allow_stage(_ctx: ValidationRequestContext) -> ValidationDecision:
    """`_allow` stub: unconditionally pass."""
    return ValidationDecision(allowed=True)
```
After:
```python
# (deleted)
```

### `backtest_simulator/launcher/action_submitter.py`::`_submit_translated` SELL fast-path
Before:
```python
if action.direction == OrderSide.SELL:
    decision = ValidationDecision(allowed=True)
    # ... fast-path bypass; logs 'CAPITAL skipped'
    return command_id, decision, context
```
After:
```python
if action.direction == OrderSide.SELL:
    decision = ValidationDecision(allowed=True)
    # ... same fast-path; log message changed to 'pipeline bypassed'.
    # Comment block above the branch now spells out the divergence:
    # (a) long-only strategy template doesn't propagate trade_id
    #     from BUY to SELL → INTAKE_TRADE_REFERENCE_INVALID;
    # (b) CapitalController has no close_position primitive.
    # Both are upstream Nexus/strategy debt; tracked at slice merge.
    return command_id, decision, context
```

### `backtest_simulator/launcher/launcher.py`::`build_validation_pipeline` call site
Before:
```python
pipeline, controller, capital_state = build_validation_pipeline(
    capital_pool=allocated_capital,
)
```
After:
```python
pipeline, controller, capital_state = build_validation_pipeline(
    nexus_config=nexus_config,
    capital_pool=allocated_capital,
)
```

---

## Tests That Prove the MVC

| MVC assertion | test file | test name |
|---|---|---|
| `_allow_stage` not in `capital.py` AST | tests/honesty/test_validation_parity.py | test_no_allow_stage_in_source |
| All 6 stages bound to non-`_allow_stage` callables | tests/honesty/test_validation_parity.py | test_all_stages_run_real_validators |
| INTAKE max_order_rate=1 denies second ENTER in 1s | tests/honesty/test_validation_parity.py | test_intake_max_order_rate_denies |
| INTAKE duplicate command_id within window denies | tests/honesty/test_validation_parity.py | test_intake_duplicate_command_id_denies |
| PRICE wide-spread snapshot denies vs `max_spread_bps` | tests/honesty/test_validation_parity.py | test_price_spread_denies |
| RISK breached drawdown limit denies | tests/honesty/test_validation_parity.py | test_risk_drawdown_denies |
| HEALTH snapshot above latency breach denies | tests/honesty/test_validation_parity.py | test_health_latency_denies |
| PLATFORM_LIMITS notional cap denies | tests/honesty/test_validation_parity.py | test_platform_notional_denies |
| Late-stage denial releases the CAPITAL reservation (no leak) | tests/honesty/test_validation_parity.py | test_late_stage_denial_releases_capital_reservation |
| SELL fast-path divergence is documented in source | tests/honesty/test_validation_parity.py | test_sell_fast_path_documented_in_source |
| INTAKE pre-hook docstrings cite parity divergence note | tests/honesty/test_validation_parity.py | test_intake_pre_hook_docstring_documents_divergence |
| Whole suite green on the merge commit | (entire suite — no single test) | n/a |
| Version 2.0.5 → 2.0.6 + CHANGELOG `# v2.0.6` line | tests/cli/test_version.py | test_pyproject_version_matches |

---

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow file: `.github/workflows/pr_checks_honesty.yml`
Job name: `pr_checks_honesty`
Specific failing command the job runs: `.venv-honesty/bin/python -m pytest tests/honesty/ -v --tb=short`

`tests/honesty/test_validation_parity.py` lives directly under `tests/honesty/` and is mechanically picked up by the command above. A future PR that re-introduces `_allow_stage`, deletes the `_resolve_validation_action` SELL→EXIT mapping, weakens any of the six stage bindings, or removes any of the per-stage denial tests fails one or more of: `test_no_allow_stage_in_source`, `test_all_stages_run_real_validators`, `test_intake_max_order_rate_denies`, `test_intake_duplicate_command_id_denies`, `test_price_spread_denies`, `test_risk_drawdown_denies`, `test_health_latency_denies`, `test_platform_notional_denies`, `test_sell_fast_path_documented_in_source`, `test_intake_pre_hook_docstring_documents_divergence` — all collected by `pytest tests/honesty/`.

A second mechanical block exists at `.github/workflows/pr_checks_lint.yml`, job `pr_checks_lint`, which runs `.venv-lint/bin/python -m pytest tests/ -q` as part of its coverage step (line 101); the same regression tests fire from the project-wide collection too. Either gate failing red blocks merge.

---

## Out of Scope

- `backtest_simulator/feed/**/*.py`
- `backtest_simulator/venue/**/*.py`
- `backtest_simulator/pipeline/**/*.py`
- `backtest_simulator/sensors/**/*.py`
- `backtest_simulator/strategies/**/*.py`
- `backtest_simulator/reporting/**/*.py`
- `backtest_simulator/cli/**/*.py`
- `backtest_simulator/exceptions.py`
- `backtest_simulator/honesty/conservation.py`
- `backtest_simulator/honesty/atr_gate.py`
- `backtest_simulator/honesty/r_per_trade.py`
- `backtest_simulator/honesty/prescient.py`
- `backtest_simulator/honesty/touch_provider.py`
- `tests/cli/**/*.py`
- `tests/launcher/**/*.py`
- `tests/pipeline/**/*.py`
- `tests/venue/**/*.py`
- `tests/integration/**/*.py`
- `tests/feed/**/*.py`
- `tests/tools/**/*.py`
- `tests/origo_source_native/**/*.py`
- `.github/workflows/**`
- `.github/rulesets/**`
- `tools/**`
- `stubs/**`

---

## Risks and Open Questions

Risks:

- Mapping SELL+ENTER → `ValidationAction.EXIT` changes which stages run on the SELL close path (currently zero; after slice five real stages, CAPITAL bypassed by Nexus's built-in EXIT-bypass logic). With MMVP-lenient defaults (all None thresholds + neutral health snapshot) every existing SELL still passes — but a future operator who dials in non-default `nexus_config.max_spread_bps` will now see SELL closes denied if the book is wide. That is correct paper-trade-parity behavior, not a regression.
- The `_check_declared_stop` / `_check_atr_sanity` pre-hooks remain bts-side. They are NOT moved into the pipeline because `ValidationRequestContext` does not carry `action.execution_params['stop_price']`. Their docstrings call this out and link the Nexus follow-up issue. The slice does not pretend to close the divergence — it documents and bounds it.

Open questions:

- (none)

---

## Closeout

> bot verifies:

Merge SHA: (to fill)
Merged PR number: (to fill)
- pr_checks_lint : (to fill)
- pr_checks_typing : (to fill)
- pr_checks_slice : (to fill)
- pr_checks_honesty : (to fill)
- pr_checks_cc : (to fill)
- pr_checks_fail_loud : (to fill)
- pr_checks_ruleset : (to fill)
- pr_checks_version : (to fill)
- pr_checks_tests : (to fill)
- PR Checks CodeQL (python) : (to fill)
- CodeQL : (to fill)

---

## Appendix — Parity Decisions

This section records the architectural decisions the template's mechanical sections do not have a slot for. It is normative: a reviewer reading this issue six months from now should be able to explain *why* the divergences below remain, and what the closure path is.

### Praxis is the parity reference

`/Users/mikkokotila/dev/Praxis/praxis/launcher.py:297-389` defines `_build_validation_pipeline(nexus_config, capital_controller, *, health/platform/price snapshot providers)` that wires every stage with MMVP-lenient defaults using exactly these helpers:

- INTAKE: `validate_intake_stage(context, build_default_intake_hooks(nexus_config))`
- RISK: `validate_risk_stage(context, RiskStageLimits())`
- PRICE: `validate_price_stage(context, build_price_stage_limits_from_config(nexus_config), price_snapshot_provider())`
- CAPITAL: `validate_capital_stage(context, capital_controller)`
- HEALTH: `validate_health_stage(context, health_snapshot_provider(), HealthStagePolicy())`
- PLATFORM_LIMITS: `validate_platform_limits_stage(context, PlatformLimitsStageLimits(), platform_snapshot_provider())`

The slice's deliverable is to mirror this shape in `backtest_simulator/honesty/capital.py`. Same code paths, same lenient defaults, same intake-hook-built-once-and-shared-across-ticks lifecycle. Operator-supplied limits are dialled in by passing a pre-configured `nexus_config` and richer snapshot providers — same knobs Praxis exposes, no new knobs invented in bts.

### SELL close path — narrowed scope decision

Approach considered: map `SELL+ENTER → ValidationAction.EXIT` so Nexus's built-in `_should_bypass_stage` skips CAPITAL/HEALTH/PLATFORM/RISK/PRICE for exits and INTAKE runs real. This IS the Praxis paper-trade shape and would close gap-C end-to-end.

Why deferred: `nexus.core.validator.intake_stage.make_reference_integrity_hook` requires `context.state.positions[trade_id]` for EXIT validation (line 230-256). Two upstream gaps stop the bts long-only flow from satisfying that:

1. **Strategy-template gap**: the long-only `BuyAndHoldStrategy` does not propagate `Action.trade_id` from the BUY leg to the matching SELL — both ride with `trade_id=None` and the action_submitter generates a fresh UUID each time. Without trade_id continuity, INTAKE_TRADE_REFERENCE_INVALID would deny every close.
2. **Capital-controller gap**: even with trade_id continuity, `CapitalController` has no `close_position` primitive. Routing SELL through CAPITAL would over-reserve.

Both fixes live upstream of bts. Opening them in this slice would balloon the diff into Nexus + strategy-template territory, breaking the single-capability rule.

What this slice ships instead: keep the SELL fast-path as documented divergence with the explicit reason in the source comment (`'pipeline bypassed'` log line + multi-line bypass-rationale comment), so a future reviewer reads "this is intentional debt with a tracked closure path" rather than "this is an unexplained `if action.direction == OrderSide.SELL:` branch". The follow-up issues (below) close gap-C properly.

### What stays bts-side as documented divergence (with closure path)

1. **`_check_declared_stop` and `_check_atr_sanity` remain in `action_submitter.py`** as bts-side pre-hooks running BEFORE `bindings.validation_pipeline.validate(context)`. The reason: `ValidationRequestContext` does not carry `action.execution_params['stop_price']`, and these checks need that field. Their docstrings rewrite to: *"BTS-side pre-hook supplementing the real Nexus INTAKE stage. The data this hook needs (`execution_params['stop_price']`) is not on `ValidationRequestContext`. Closure path: Nexus PR extending `ValidationRequestContext.execution_params` (filed as a follow-up Nexus issue at slice merge)."* The hook lives where it does because the data lives where it does — it is not a stub, not a stage replacement, not a workaround for `_allow`. With this slice INTAKE itself is real.

2. **`CapitalController` has no `close_position` primitive.** Nexus's pipeline knows about EXIT-bypass for CAPITAL, but the matching close-side accounting still happens via the bts `CapitalLifecycleTracker.record_close_position` path (which is correct — it mirrors `CapitalController.order_fill` exactly: releases `cost_basis + entry_fees` from `position_notional`, decrements `per_strategy_deployed`). Closure path: a future Nexus PR adding `CapitalController.close_position(reservation_id, fill_notional, fees)` lets the SELL fast-path round-trip through CAPITAL too. Tracked as follow-up.

### Carried-forward debt (filed at slice merge)

- New issue: *"Extend Nexus `ValidationRequestContext` with `execution_params` so bts INTAKE pre-hooks collapse into the pipeline."* Closes the `_check_declared_stop` / `_check_atr_sanity` divergence end-to-end.
- New issue: *"Long-only strategy template must propagate `Action.trade_id` from BUY to SELL."* Necessary precondition for SELL→EXIT mapping to work end-to-end through Nexus's INTAKE reference-integrity check.
- New issue: *"Add `CapitalController.close_position` primitive in Nexus."* Closes the SELL-close-of-a-long CAPITAL bypass — the EXIT route would then run all six stages including a real CAPITAL release.

All three are upstream Nexus / strategy-template changes, out of bts slice scope. The Nexus context-extension and the CapitalController primitive are listed under *Routing → Unlocks* above. Together they unlock a follow-up bts slice that maps SELL→EXIT and runs every stage real on every action.

mikkokotila:
## Summary

Closes #28. Replaces five `_allow_stage` stubs in `backtest_simulator/honesty/capital.py::build_validation_pipeline` with real Nexus `validate_*_stage` calls, mirroring the Praxis paper-trade reference at `praxis/launcher.py::_build_validation_pipeline`. MMVP-lenient defaults preserve existing test pass-through; operator-supplied `nexus_config` / limits / snapshot providers dial in real denial behavior.

- **INTAKE / RISK / PRICE / HEALTH / PLATFORM_LIMITS** now run real validators (CAPITAL was already real).
- **Late-stage denial leak fixed**: HEALTH/PLATFORM denials carry the CAPITAL reservation forward; `_submit_translated` now releases it via `bindings.capital_controller.release_reservation` so available capital is not silently leaked.
- **SELL fast-path stays bypassed** as documented divergence with two upstream debt items (long-only strategy template doesn't propagate `Action.trade_id`; `CapitalController` has no `close_position` primitive).
- **Pre-hooks remain bts-side** (`_check_declared_stop`, `_check_atr_sanity`) — `ValidationRequestContext` does not carry `execution_params['stop_price']`. Docstrings rewritten to cite the parity-divergence phrase + closure path.

11 tests in `tests/honesty/test_validation_parity.py` lock the contract.

## Budget raises

[budget-raise: backtest_simulator/honesty/capital.py: real Nexus stage closures + lenient-default factories + parity-mirroring snapshot providers — Praxis equivalent runs ~150 lines]
[budget-raise: backtest_simulator/launcher/action_submitter.py: SubmitterBindings gains capital_controller; denied-decision release path adds ~12 lines; pre-hook docstrings carry parity-divergence note]

## Test plan

- [x] `pytest tests/honesty/test_validation_parity.py -q` — 11 passed
- [x] `pytest -q` — 426 passed
- [x] `python -m backtest_simulator.cli gate lint` — All checks passed!
- [x] All 4 bloat gates pass locally
- [x] codex post-auditor-1 P1 (capital reservation leak) addressed
- [ ] codex post-auditor-2 verdict (in flight)
- [ ] zero-bang review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ Two parity gaps remain explicit-in-source rather than stubbed:
 
 ## Test surface
 
-`tests/honesty/test_validation_parity.py` adds 10 tests:
+`tests/honesty/test_validation_parity.py` adds 11 tests:
 - `_allow_stage` AST-absent
 - All 6 stages bound to non-`_allow_stage` callables
 - Per-stage end-to-end denials (INTAKE × 2, PRICE, RISK, HEALTH,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,55 @@
+# v2.0.6
+
+Validator-parity slice — five Nexus pipeline stages that were `_allow`
+stubs now run real `validate_*_stage` calls, mirroring Praxis paper-
+trade's `_build_validation_pipeline` with MMVP-lenient defaults
+(`RiskStageLimits()`, `PlatformLimitsStageLimits()`, `HealthStagePolicy()`,
+`PriceStageLimits` derived from `nexus_config`). Operator-supplied
+limits dial in real denial behavior — same dials Praxis exposes, no
+bts-specific knobs invented.
+
+## Real validators wired
+
+`backtest_simulator/honesty/capital.py::build_validation_pipeline`
+deletes `_allow_stage` and binds:
+
+- INTAKE → `validate_intake_stage(context, build_default_intake_hooks(nexus_config))`
+- RISK → `validate_risk_stage(context, RiskStageLimits())`
+- PRICE → `validate_price_stage(context, build_price_stage_limits_from_config(nexus_config), price_snapshot_provider())`
+- CAPITAL → unchanged (already real)
+- HEALTH → `validate_health_stage(context, health_snapshot_provider(), HealthStagePolicy())`
+- PLATFORM_LIMITS → `validate_platform_limits_stage(context, PlatformLimitsStageLimits(), platform_snapshot_provider())`
+
+The signature gains `nexus_config: InstanceConfig` (required), plus
+keyword-only `risk_limits`, `health_policy`, `platform_limits`, and
+three snapshot providers — same shape as Praxis.
+
+## Documented divergence (carried forward)
+
+Two parity gaps remain explicit-in-source rather than stubbed:
+
+1. `_check_declared_stop` and `_check_atr_sanity` continue to run as
+   bts-side INTAKE pre-hooks because `ValidationRequestContext` does
+   not expose `action.execution_params['stop_price']`. Their
+   docstrings now cite the closure path (Nexus PR extending the
+   context).
+2. The SELL close fast-path in `action_submitter._submit_translated`
+   stays in place because (a) the long-only strategy template does
+   not propagate `Action.trade_id` from BUY to SELL, and (b)
+   `CapitalController` has no `close_position` primitive. Both
+   resolutions are upstream Nexus/strategy work; the source comment
+   spells out the debt explicitly.
+
+## Test surface
+
+`tests/honesty/test_validation_parity.py` adds 10 tests:
+- `_allow_stage` AST-absent
+- All 6 stages bound to non-`_allow_stage` callables
+- Per-stage end-to-end denials (INTAKE × 2, PRICE, RISK, HEALTH,
+  PLATFORM_LIMITS) via `nexus_config` / limits / snapshot providers
+- INTAKE pre-hook docstrings cite parity divergence
+- SELL fast-path divergence documented in source
+
 # v2.0.5
 
 Package-cleanup slice — no functional change. Drops operator-local
--- a/backtest_simulator/launcher/action_submitter.py
+++ b/backtest_simulator/launcher/action_submitter.py
@@ -294,6 +294,7 @@ class SubmitterBindings:
     state: InstanceState
     praxis_outbound: PraxisOutbound
     validation_pipeline: ValidationPipeline
+    strategy_budget: Decimal
     # Slice #28 — pipeline.validate's CAPITAL stage runs before
     # HEALTH/PLATFORM_LIMITS, so a denial from a later stage carries
     # the reservation forward (per nexus pipeline_executor). Without
@@ -304,7 +305,6 @@ class SubmitterBindings:
     # state. Default-None for tests that don't care (a denied
     # decision with reservation=None is a no-op release path).
     capital_controller: CapitalController | None = None
-    strategy_budget: Decimal = Decimal('0')
     touch_provider: Callable[[str], Decimal | None] | None = None
     tick_provider: Callable[[str], Decimal] | None = None
     # Slice #17 Task 29 — ATR R-denominator gameability gate.
--- a/backtest_simulator/launcher/action_submitter.py
+++ b/backtest_simulator/launcher/action_submitter.py
@@ -520,13 +520,23 @@ def _submit_translated(
                 decision.reservation.reservation_id,
             )
             if not release_result.success:
-                _log.warning(
-                    'pipeline-denied reservation release failed: '
-                    'reservation_id=%s reason=%s category=%s',
-                    decision.reservation.reservation_id,
-                    release_result.reason,
-                    release_result.category,
+                # Fail loud — the reservation was just minted by the same
+                # pipeline.validate() call on the same controller; any
+                # non-success here is a wiring/state bug (controller
+                # mismatch, concurrent release, lost reservation), not an
+                # expected race. Mirrors `CapitalLifecycleTracker.record_rejection`'s
+                # pattern at honesty/capital.py:461. Letting the bug
+                # through with a warning would silently re-introduce
+                # the very leak this branch exists to prevent
+                # (bit-mis no-defects-conviction P1).
+                msg = (
+                    f'pipeline-denied reservation release failed: '
+                    f'reservation_id={decision.reservation.reservation_id} '
+                    f'failed_stage={decision.failed_stage} '
+                    f'reason={release_result.reason!r} '
+                    f'category={release_result.category}'
                 )
+                raise RuntimeError(msg)
         _log.warning(
             'validation denied: stage=%s reason_code=%s message=%s command_id=%s',
             decision.failed_stage.value if decision.failed_stage else None,
--- a/backtest_simulator/launcher/action_submitter.py
+++ b/backtest_simulator/launcher/action_submitter.py
@@ -12,6 +12,7 @@ from enum import Enum
 from types import SimpleNamespace
 from typing import cast
 
+from nexus.core.capital_controller.capital_controller import CapitalController
 from nexus.core.domain.enums import OrderSide
 from nexus.core.domain.instance_state import InstanceState
 from nexus.core.domain.order_types import ExecutionMode as _NexusExecutionMode
@@ -293,7 +294,17 @@ class SubmitterBindings:
     state: InstanceState
     praxis_outbound: PraxisOutbound
     validation_pipeline: ValidationPipeline
-    strategy_budget: Decimal
+    # Slice #28 — pipeline.validate's CAPITAL stage runs before
+    # HEALTH/PLATFORM_LIMITS, so a denial from a later stage carries
+    # the reservation forward (per nexus pipeline_executor). Without
+    # explicit release, available capital leaks. The action_submitter
+    # uses this controller to release reservations attached to denied
+    # decisions; the launcher constructs both pipeline + controller
+    # from the same `build_validation_pipeline` call so they share
+    # state. Default-None for tests that don't care (a denied
+    # decision with reservation=None is a no-op release path).
+    capital_controller: CapitalController | None = None
+    strategy_budget: Decimal = Decimal('0')
     touch_provider: Callable[[str], Decimal | None] | None = None
     tick_provider: Callable[[str], Decimal] | None = None
     # Slice #17 Task 29 — ATR R-denominator gameability gate.
@@ -337,10 +348,16 @@ def build_action_submitter(
         calls can tie back to the original reservation.
       - `on_submit` is the pre-existing drain hook the clock loop uses.
 
-    The only stage this pipeline actually enforces in Part 2 is CAPITAL;
-    INTAKE / RISK / PRICE / HEALTH / PLATFORM_LIMITS are `_allow` stubs
-    as documented in `backtest_simulator.honesty.capital`. ABORT flows
-    through `praxis_outbound.send_abort` unchanged.
+    Slice #28: every Nexus pipeline stage runs a real `validate_*_stage`
+    call. INTAKE / RISK / PRICE / CAPITAL / HEALTH / PLATFORM_LIMITS are
+    all backed by their respective Nexus validators with MMVP-lenient
+    defaults; operator-supplied limits + snapshot providers in
+    `backtest_simulator.honesty.capital.build_validation_pipeline` dial
+    in real denial behavior. CAPITAL runs before HEALTH/PLATFORM_LIMITS
+    in stage order, so a late-stage denial that carries the reservation
+    forward triggers an explicit `capital_controller.release_reservation`
+    in `_submit_translated`. ABORT flows through `praxis_outbound.send_abort`
+    unchanged.
     """
     def _submit(actions: list[Action], strategy_id: str) -> None:
         for raw_action in actions:
@@ -492,6 +509,24 @@ def _submit_translated(
         return command_id, decision, context
     decision = bindings.validation_pipeline.validate(context)
     if not decision.allowed:
+        # Slice #28: CAPITAL is stage 4 of 6, so HEALTH/PLATFORM_LIMITS
+        # denials carry the CAPITAL-stage reservation forward in the
+        # denied decision (nexus pipeline_executor preserves it). The
+        # action is dropped without reaching Praxis, so no fill / no
+        # `order_ack` / no `order_fill` will retire the reservation.
+        # Release it explicitly so available capital is not leaked.
+        if decision.reservation is not None and bindings.capital_controller is not None:
+            release_result = bindings.capital_controller.release_reservation(
+                decision.reservation.reservation_id,
+            )
+            if not release_result.success:
+                _log.warning(
+                    'pipeline-denied reservation release failed: '
+                    'reservation_id=%s reason=%s category=%s',
+                    decision.reservation.reservation_id,
+                    release_result.reason,
+                    release_result.category,
+                )
         _log.warning(
             'validation denied: stage=%s reason_code=%s message=%s command_id=%s',
             decision.failed_stage.value if decision.failed_stage else None,
--- a/backtest_simulator/launcher/action_submitter.py
+++ b/backtest_simulator/launcher/action_submitter.py
@@ -294,7 +294,6 @@ class SubmitterBindings:
     state: InstanceState
     praxis_outbound: PraxisOutbound
     validation_pipeline: ValidationPipeline
-    strategy_budget: Decimal
     # Slice #28 — pipeline.validate's CAPITAL stage runs before
     # HEALTH/PLATFORM_LIMITS, so a denial from a later stage carries
     # the reservation forward (per nexus pipeline_executor). Without
@@ -302,9 +301,10 @@ class SubmitterBindings:
     # uses this controller to release reservations attached to denied
     # decisions; the launcher constructs both pipeline + controller
     # from the same `build_validation_pipeline` call so they share
-    # state. Default-None for tests that don't care (a denied
-    # decision with reservation=None is a no-op release path).
-    capital_controller: CapitalController | None = None
+    # state. REQUIRED — making it optional would let a caller silently
+    # regress to the leak shape (slice-#28 audit, finding 5).
+    capital_controller: CapitalController
+    strategy_budget: Decimal
     touch_provider: Callable[[str], Decimal | None] | None = None
     tick_provider: Callable[[str], Decimal] | None = None
     # Slice #17 Task 29 — ATR R-denominator gameability gate.
@@ -515,7 +515,7 @@ def _submit_translated(
         # action is dropped without reaching Praxis, so no fill / no
         # `order_ack` / no `order_fill` will retire the reservation.
         # Release it explicitly so available capital is not leaked.
-        if decision.reservation is not None and bindings.capital_controller is not None:
+        if decision.reservation is not None:
             release_result = bindings.capital_controller.release_reservation(
                 decision.reservation.reservation_id,
             )
--- a/backtest_simulator/launcher/action_submitter.py
+++ b/backtest_simulator/launcher/action_submitter.py
@@ -114,6 +114,8 @@ _ACTION_TYPE_TO_VALIDATION_ACTION: dict[ActionType, ValidationAction] = {
 }
 
 
+
+
 # --- Nexus -> Praxis enum converters ----------------------------------------
 # Nexus and Praxis define separate Enum classes for the same domain
 # concepts (ExecutionMode, OrderType, OrderSide, MakerPreference,
@@ -430,14 +432,14 @@ def _submit_translated(
         strategy_id=strategy_id, action=action,
         strategy_budget=bindings.strategy_budget,
     )
-    # Part 2 INTAKE hook: declared-stop enforcement. Long-only strategy
-    # convention — BUY opens a long (requires protective stop), SELL
-    # closes a long (is itself the exit, no new stop). The hook runs
-    # BEFORE `validation_pipeline.validate` because the pipeline's
-    # INTAKE stage is `_allow` in Part 2 and the stop is carried on
-    # `action.execution_params`, which isn't accessible from within a
-    # `StageValidator`. Slice #17 Task 29 layers the ATR sanity check
-    # on top: stop must be ≥ `gate.k * ATR(window)` from entry.
+    # BTS-side INTAKE pre-hook supplementing the real Nexus INTAKE stage.
+    # Slice #28 lit up the pipeline's INTAKE with real
+    # `validate_intake_stage(...)` checks; these two hooks layer on top
+    # because the data they need (`execution_params['stop_price']`) is
+    # not on `ValidationRequestContext`. Closure path: a Nexus PR
+    # extending `ValidationRequestContext.execution_params`. Long-only
+    # strategy convention — BUY opens a long (requires protective stop),
+    # SELL closes a long (is itself the exit, no new stop).
     intake_decision = _check_declared_stop(action, context)
     if intake_decision is None:
         intake_decision = _check_atr_sanity(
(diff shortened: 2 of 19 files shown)


backtest_simulator/launcher/action_submitter.py
```diff
@@ -291,7 +294,17 @@ class SubmitterBindings:
...
+    # HEALTH/PLATFORM_LIMITS, so a denial from a later stage carries
+    # the reservation forward (per nexus pipeline_executor). Without
+    # explicit release, available capital leaks. The action_submitter
+    # uses this controller to release reservations attached to denied
+    # decisions; the launcher constructs both pipeline + controller
+    # from the same `build_validation_pipeline` call so they share
+    # state. Default-None for tests that don't care (a denied
+    # decision with reservation=None is a no-op release path).
+    capital_controller: CapitalController | None = None
+    strategy_budget: Decimal = Decimal('0')
```
zero-bang: `strategy_budget` was previously required (no default). Adding `capital_controller` before it forced a `Decimal('0')` default to satisfy Python's dataclass field-ordering rule. This silently weakens the API — a future caller that omits `strategy_budget` gets CAPITAL denials with no compile-time or type-check error.

Fix: swap the two fields so `strategy_budget` stays required:

```python
    strategy_budget: Decimal
    capital_controller: CapitalController | None = None
```

All existing callers use keyword args, so the reorder is safe. The launcher in `launcher.py:1160` already passes both explicitly.

tests/honesty/test_validation_parity.py
```diff
@@ -0,0 +1,339 @@
...
+        nexus_config=cfg, capital_pool=Decimal('100000'),
+        health_policy=policy, health_snapshot_provider=hot_snapshot,
+    )
+    ctx = _ctx(
+        command_id='cmd-leak-1', nexus_config=cfg,
+        capital_state=capital_state,
+    )
+    decision = pipeline.validate(ctx)
+    # The pipeline returns a denied decision carrying the CAPITAL
+    # reservation forward. The action_submitter will release it; we
```
zero-bang: This test proves the *components* work (pipeline produces a reservation on late-stage denial, `release_reservation` retires it) but doesn't exercise the *integration point* where the P1 fix lives: `_submit_translated`'s `if decision.reservation is not None and bindings.capital_controller is not None:` branch. If that branch were deleted, this test would still pass.

Worth adding a companion test that drives a BUY through `_submit_translated` with a health-denied pipeline and `capital_controller` set, then asserts `capital_state.reservation_notional == 0` after the call returns `None`. That locks the release path end-to-end rather than only the underlying primitives. Not blocking merge, but the P1 fix deserves a test that would break if the fix were removed.

CHANGELOG.md
```diff
@@ -1,3 +1,55 @@
...
+2. The SELL close fast-path in `action_submitter._submit_translated`
+   stays in place because (a) the long-only strategy template does
+   not propagate `Action.trade_id` from BUY to SELL, and (b)
+   `CapitalController` has no `close_position` primitive. Both
+   resolutions are upstream Nexus/strategy work; the source comment
+   spells out the debt explicitly.
+
+## Test surface
+
+`tests/honesty/test_validation_parity.py` adds 10 tests:
```
zero-bang: Says "adds 10 tests" but `test_validation_parity.py` defines 11 test functions. Minor — update the count.

zero-bang (changes requested):
Solid slice — the wiring mirrors Praxis faithfully, the lenient defaults are the right posture, the late-stage reservation leak fix is the right shape, and the test surface is thorough. One blocking field-ordering issue on `SubmitterBindings`, one test-coverage note.

Forward-looking concern: the next predictable failure mode is an operator who supplies a non-default `max_spread_bps` via `nexus_config` and is surprised that BUY actions are denied on PRICE during backtests where the historical tape has no spread data — `_default_price_snapshot` returns `None` which passes the lenient default, but a future snapshot provider wired to the tape could surface stale or missing spread data as `spread_bps=Decimal('inf')` and deny everything silently.

backtest_simulator/launcher/action_submitter.py
mikkokotila: Fixed in 6e43109 — `strategy_budget` is back to required (no default), `capital_controller: CapitalController | None = None` follows it. All 426 tests still green, lint clean.

tests/honesty/test_validation_parity.py
mikkokotila: Already addressed at 20af2a1 (one commit later than the one you reviewed): the test now builds `SubmitterBindings(capital_controller=controller, ...)`, drives `submit([action], "long_on_signal")` through `build_action_submitter`, and asserts both `outbound.commands == 0` and `capital_state.reservation_notional == 0`. Mutation-verified: changing `if decision.reservation is not None and bindings.capital_controller is not None:` to `if False and ...` in `_submit_translated` makes the test fail with `reservation_notional=53.60700000`. Codex post-auditor-3 ran the same mutation and approved.

CHANGELOG.md
mikkokotila: Fixed in 6e43109 — `# v2.0.6` line now reads "adds 11 tests".

mikkokotila:
All review threads addressed:

1. **SubmitterBindings field ordering** — fixed in 6e43109. `strategy_budget` is required again (no default); `capital_controller: CapitalController | None = None` follows it.
2. **Test-coverage on `_submit_translated`** — already addressed in 20af2a1 (one commit after the one you reviewed). The test now drives `submit([action], 'long_on_signal')` end-to-end and asserts both `outbound.commands == 0` and `capital_state.reservation_notional == 0`. Mutation-verified: changing the production guard at `action_submitter.py:518` to `if False ...` fails the test with `reservation_notional=53.60700000`. Codex post-auditor-3 ran the same mutation independently and approved.
3. **CHANGELOG test count** — fixed in 6e43109 (10 → 11).
4. **Forward-looking concern about tape-backed price snapshots** — `_default_price_snapshot`'s docstring at d1055a4 now spells out the closure path: a future provider must surface missing data as `PRICE_SYSTEM_DATA_UNAVAILABLE` (already implemented in Nexus), never paper over it with `Decimal('inf')`.

426 tests still pass; lint clean; all 10 CI gates green. Re-requesting your review.

zero-bang (approved):

mikkokotila:
Operator audit finding 5 addressed in 7df5dc1: `capital_controller` is now required on `SubmitterBindings` (no default). Any caller that omits it gets a TypeError at construction, not a silent reservation leak. The 8 test fixtures that previously discarded `_controller` now capture and thread it through. Mutation re-verified: disabling the production release block fails the leak-release test.

Filed follow-ups for the other audit findings (out of slice scope, but worth tracking):
- #30 — `bts gate lint` still references deleted `scripts/` path (BTS-tooling-trust)
- #31 — local `bts gate typecheck` diverges from CI `pr_checks_typing` (BTS-tooling-trust)
- #32 — `bts run/sweep` reuses stale Limen experiment output without binding cache key to code SHA (P0 market-relevance, 'strategy tested is strategy deployed')

Re-requesting your review on the finding-5 fix.

zero-bang (approved):

bit-mis:
### bit-mis review — Concrete Value for the User

**P1**
- The PR title's "parity with paper-trade" claim is structurally true but operationally inert at the only user-facing entry point. `BacktestLauncher` calls `build_validation_pipeline(nexus_config=nexus_config, capital_pool=allocated_capital)` with zero overrides — `backtest_simulator/launcher/launcher.py:1128-1131` — and `_build_nexus_instance_config` constructs `NexusInstanceConfig(account_id=..., venue='binance_spot_simulated')` with no `max_order_rate`, `duplicate_order_window_seconds`, or `max_spread_bps` — `backtest_simulator/launcher/launcher.py:1256-1261`. With every newly-wired stage at its lenient default (`_default_health_snapshot` fully healthy at `capital.py:60-74`; `_default_platform_snapshot` empty at `capital.py:77-79`; `_default_price_snapshot` returns `None` at `capital.py:82-95`; `RiskStageLimits()`/`HealthStagePolicy()`/`PlatformLimitsStageLimits()` all-threshold-None at `capital.py:134-141`), no `bts run` invocation can possibly fire an INTAKE/RISK/PRICE/HEALTH/PLATFORM_LIMITS denial. CAPITAL (already real on `main`) is the sole stage that denies anything. The user's observable backtest behavior — fills, denials, PnL — is identical pre- and post-merge.
- The late-stage leak fix at `backtest_simulator/launcher/action_submitter.py:509-527` only fires when HEALTH/PLATFORM_LIMITS deny after CAPITAL has reserved. Because the launcher cannot construct a config that makes those stages deny (per finding above), the fix delivers no observable value to default users — it is exercised only by `tests/honesty/test_validation_parity.py::test_late_stage_denial_releases_capital_reservation`, which manually injects a `HealthStagePolicy` and hot snapshot the launcher has no path to produce. The PR body's "available capital is not silently leaked" reads as a present-tense user benefit; in practice it is dormant code waiting on an operator-configuration surface this slice does not add.

bit-mis:
### bit-mis review — No Defects Conviction

**P1**
- Pipeline-denied reservation release silently swallows controller failures — `backtest_simulator/launcher/action_submitter.py:522` — when `release_reservation` returns `success=False` the code only emits `_log.warning` and continues. The reservation being released was just minted by `validate_capital_stage` in the same `pipeline.validate()` call and the controller is the same instance the launcher constructs alongside the pipeline (`launcher.py:1126-1162`), so any non-success here means a wiring/state bug (controller mismatch, concurrent release, lost reservation), not an expected race. The sibling lifecycle path `CapitalLifecycleTracker.record_rejection` raises `RuntimeError` on the identical `LifecycleResult` (capital.py:461), and the project's fail-loud rule (CLAUDE.md "no silent-fallback patterns") treats warning-then-continue as a hide. The very leak this block exists to prevent will silently re-appear if `release_result.success` flips to `False`, while the only signal is a log line buried among normal denials. Mirror `record_rejection` and raise on failure (or at minimum on any non-`EXPECTED_MISS` category).

mikkokotila:
@bit-mis — addressed both findings.

**No-defects-conviction P1 (silent release failure)** — fixed in 37370e8. `_submit_translated` now raises `RuntimeError` with reservation_id + failed_stage + reason + category when `release_reservation` returns success=False, mirroring `CapitalLifecycleTracker.record_rejection` (capital.py:461). New test `tests/honesty/test_validation_parity.py::test_release_reservation_failure_fails_loud` proves the fail-loud path: it wraps the controller so release returns success=False and asserts RuntimeError propagates through `submit`. 12 parity tests pass; 427 in the full suite.

**Concrete-value P1 (operationally inert parity)** — your read is right: with the launcher passing zero overrides, every newly-wired stage stays at its lenient default, so no `bts run` invocation produces a non-CAPITAL denial. The slice's MVC predicates are met (wiring exists, no `_allow_stage`, leak path is structurally correct), but the user-visible behavior is identical pre/post-merge. That's a real paper-parity gap.

It's separate scope from "replace the stubs". I've filed it as **#33** — *feat(launcher): expose validator-stage knobs so bts run produces non-CAPITAL denials* — with the file:line citations from your review and a 4-surface scope plan: manifest schema extension → `_build_nexus_instance_config` consuming it → `build_validation_pipeline` wiring it → tape-backed snapshot providers (phase 2). The release-leak fix from this slice becomes operationally exercised once #33 lands.

If you'd rather see #33's surface folded into this slice instead, say the word — I'll expand. Otherwise this PR closes A+B (real validators) and #33 closes the operator-observability layer in its own slice.

bit-mis (approved):
bit-mis: all green; no unresolved P0/P1 from facet review or audit. Approving.

Issue #65, opened 2026-05-06
RFC-2001: Deterministic bts sweep via schedule-driven replay clock


mikkokotila:
## Objective

`bts sweep` is non-deterministic today. Two back-to-back invocations of the same canonical command on the same bundle and replay window produce different per-window CSVs, different trade counts, and different P&L for the same `(decoder, day)` pair. The cause is a real race inside one subprocess between freezegun-pinned `datetime.now(UTC)` reads in a `threading.Timer` poll and `freezer.tick(...)` writes in the launcher's main thread. Same model probabilities surface at different reported timestamps; some kline boundaries are missed; some force-flatten / window-edge calls bleed into the per-tick CSV. The "strategy tested = strategy deployed" promise of `limen → nexus → praxis` is silently violated: the simulator's predict-loop schedule does not match what production Praxis fires.

We must make `bts sweep` deterministic.

## Summary

Introduce a new replay-clock module (`backtest_simulator/launcher/replay_clock.py`) that drives one window through a deterministic schedule using Nexus's freshly merged public `PredictLoop.tick_once(wired)` (Vaquum/Nexus#61, `vaquum-nexus 0.41.0`). The launcher precomputes every kline boundary inside the trading window, jumps frozen time to each boundary, calls `tick_once` synchronously per wired sensor, and drains submits + outcomes to quiescence before advancing. The legacy `threading.Timer.run` monkey-patch is removed; the synchronous replay path no longer starts the Timer-driven `PredictLoop` or the worker-thread `OutcomeLoop`. Production Praxis is unchanged.

## Is This Really Necessary?

Yes, before anything else in `The-Plan.md`. The cleanup plan's mechanical floor — *normalized canonical `bts sweep` outputs are byte-equal across two runs* — is unreachable while `bts sweep` is non-deterministic, and the cleanup plan's "every kept line is hit by canonical `bts sweep` coverage" gate is meaningless if the canonical run can't be reproduced. Until this is fixed, every result the simulator has ever produced is suspect by construction: P&L, Kelly, CPCV PBO, parity-passed runs all depended on a clock whose firing instants drifted under load.

The alternatives we considered (see *Alternatives*) reduce flake but do not deliver determinism, so they only delay the same work and pollute the cleanup with non-load-bearing mitigations.

## Named Concepts

[...]

PR #66, opened 2026-05-06
feat(launcher): replace timer clock with replay tick_once driver

### LINKED ISSUE #64: feat(launcher): replace timer clock with replay tick_once driver
## Background — read this first (out-of-template)

This slice is a **prerequisite** to the package cleanup plan in `The-Plan.md`. It does not belong to that plan's slices 1-7. The plan's mechanical floor — "normalized canonical `bts sweep` outputs are byte-equal across two runs" — is unreachable today because `bts sweep` is non-deterministic. Two back-to-back invocations of the same canonical command on the same bundle and replay window produce different per-window CSVs, different trade counts, and different P&L for the same `(decoder, day)` pair.

The root cause is a real race inside one subprocess. `backtest_simulator/launcher/clock.py` monkey-patches `threading.Timer.run` to poll `datetime.now(UTC)` (frozen by freezegun) every 5 ms. Meanwhile `backtest_simulator/launcher/launcher.py:_advance_clock_until` calls `freezer.tick(...)` to advance the same frozen clock in big jumps. The two threads share the clock with no coordination — the launcher's main thread can advance frozen time past a Timer's target between two of the Timer thread's 5 ms polls, so when the Timer thread next gets the CPU, the firing instant has already drifted by an arbitrary amount. Captured `signal.timestamp` lands on the wrong frozen instant, the strategy's number of `produce_signal` firings per day varies, and trade counts diverge. The launcher comment at `launcher.py:1713-1717` admits this: *"frozen time drifts because multiple threads (main + per-Timer conditional-sleep) tick it concurrently."*

A massive package cleanup follows immediately after (The-Plan.md slices 1-7: package LOC must drop from ~17 k to <= 3 k; every surviving line proven by canonical sweep coverage; dead files deleted). Whatever clock infrastructure this slice introduces will live through that cleanup. So this slice does not retrofit the racy clock — it introduces a new replay-clock module (`backtest_simulator/launcher/replay_clock.py`) and routes the launcher through it. The threading.Timer monkey-patch is removed. After the cleanup, `replay_clock.py` is one of the survivors; `clock.py` is reduced to whatever residual freezegun wrapper still has callers, or absorbed in a later slice.

The new module uses Nexus's freshly-merged public `PredictLoop.tick_once(wired)` ([Vaquum/Nexus#61](https://github.com/Vaquum/Nexus/pull/61), `vaquum-nexus 0.41.0`, commit `cf27f5a`). Same chain Praxis runs in production, just driven from a deterministic schedule instead of a wall-clock `threading.Timer`. The simulator stops faking real-time and instead jumps frozen time to each pre-computed kline boundary, calls `tick_once` synchronously, and drains submits + outcomes to quiescence before advancing.

**Hard invariant for the synchronous replay path.** The new path must NOT call `predict_loop.start()` or `outcome_loop.start()`. Both are background-thread starters; both make `tick_once` raise `RuntimeError` if invoked while running. The replay path drives both loops synchronously via their public single-shot entry points (`PredictLoop.tick_once`, `OutcomeLoop.tick_once`).

**Out-of-scope corollaries** (named explicitly so they don't smuggle in): no extension of TimerLoop / strategy-authored `on_timer` to the replay path (sequencer with non-empty `timer_specs` fails loud); no asyncio loop wiring changes; no changes to `_make_outcome_router`'s router; no changes to the venue adapter, account loop, or fill chain.

---

## Routing

Parent issue: #65
Depends on:
- Vaquum/Nexus#61
Unlocks:
-

---

## Minimum Viable Capability

>
> Accepted:
>
> Rejected:
> - *Strict CQRS separation.*

A. Module structure (ast-based, robust to `from __future__ import annotations`):

- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/replay_clock.py').read_text()); names={n.name for n in ast.walk(tree) if isinstance(n, ast.FunctionDef)} | {n.name for n in ast.walk(tree) if isinstance(n, ast.ClassDef)}; print('compute_kline_boundaries' in names and 'ReplayClock' in names)"` prints `True`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/replay_clock.py').read_text()); fn=next(n for n in ast.walk(tree) if isinstance(n, ast.FunctionDef) and n.name=='compute_kline_boundaries'); kwonly=[a.arg for a in fn.args.kwonlyargs]; pos=[a.arg for a in fn.args.args]; print(pos==[] and kwonly==['window_start','window_end','interval_seconds'])"` prints `True`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/replay_clock.py').read_text()); cls=next(n for n in ast.walk(tree) if isinstance(n, ast.ClassDef) and n.name=='ReplayClock'); m=next(n for n in cls.body if isinstance(n, ast.FunctionDef) and n.name=='drive_window'); print([a.arg for a in m.args.kwonlyargs]==['window_start','window_end','wired_sensors','predict_loop','outcome_loop','drain_pending_submits','freezer'])"` prints `True`.

B. Old surfaces removed (ast-based, both call-site forms):

- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/clock.py').read_text()); print(any(isinstance(n, ast.FunctionDef) and n.name=='_frozen_aware_timer_run' for n in ast.walk(tree)))"` prints `False`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/clock.py').read_text()); patches=[n for n in ast.walk(tree) if (isinstance(n, ast.Call) and isinstance(n.func, ast.Name) and n.func.id=='setattr' and any(isinstance(a, ast.Attribute) and a.attr=='Timer' for a in n.args)) or (isinstance(n, ast.Assign) and any(isinstance(t, ast.Attribute) and t.attr=='run' and isinstance(t.value, ast.Attribute) and t.value.attr=='Timer' for t in n.targets))]; print(len(patches))"` prints `0`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); print(any(isinstance(n, ast.FunctionDef) and n.name=='_advance_clock_until' for n in ast.walk(tree)))"` prints `False`.

C. Launcher *uses* the new module (not just imports):

- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); imports=[n for n in ast.walk(tree) if isinstance(n, ast.ImportFrom) and n.module=='backtest_simulator.launcher.replay_clock']; print(len(imports)>=1)"` prints `True`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); cls=next(n for n in ast.walk(tree) if isinstance(n, ast.ClassDef) and n.name=='BacktestLauncher'); rw=next(n for n in ast.walk(cls) if isinstance(n, ast.FunctionDef) and n.name=='run_window'); calls=[n for n in ast.walk(rw) if isinstance(n, ast.Call) and isinstance(n.func, ast.Attribute) and n.func.attr=='drive_window']; print(len(calls)>=1)"` prints `True`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); cls=next(n for n in ast.walk(tree) if isinstance(n, ast.ClassDef) and n.name=='BacktestLauncher'); rw=next(n for n in ast.walk(cls) if isinstance(n, ast.FunctionDef) and n.name=='run_window'); call=next(n for n in ast.walk(rw) if isinstance(n, ast.Call) and isinstance(n.func, ast.Attribute) and n.func.attr=='drive_window'); print(set(k.arg for k in call.keywords) >= {'window_start','window_end','wired_sensors','predict_loop','outcome_loop','drain_pending_submits','freezer'})"` prints `True`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); calls=[n for n in ast.walk(tree) if isinstance(n, ast.Call) and isinstance(n.func, ast.Attribute) and n.func.attr=='start' and isinstance(n.func.value, ast.Name) and n.func.value.id in {'predict_loop','outcome_loop'}]; print(len(calls))"` prints `0`.

D. TimerLoop fail-loud (the synchronous replay path must not start any background loop):

- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); calls=[n for n in ast.walk(tree) if isinstance(n, ast.Call) and isinstance(n.func, ast.Name) and n.func.id=='TimerLoop']; print(len(calls))"` prints `0`.
- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); raises=[n for n in ast.walk(tree) if isinstance(n, ast.Raise) and isinstance(n.exc, ast.Call) and isinstance(n.exc.func, ast.Name) and n.exc.func.id=='RuntimeError' and any(isinstance(c, ast.Constant) and isinstance(c.value, str) and 'timer_specs' in c.value for c in ast.walk(n.exc))]; print(len(raises)>=1)"` prints `True` (walks into the RuntimeError call, including any f-string `JoinedStr` fragments, so the AST check tolerates the f-string-inlined raise without requiring a flat `Constant` arg).

E. PraxisInbound poll_timeout=0.0 in the replay path (ast-based, multi-line tolerant):

- `python -c "import ast, pathlib; tree=ast.parse(pathlib.Path('backtest_simulator/launcher/launcher.py').read_text()); kw=[k for n in ast.walk(tree) if isinstance(n, ast.Call) and isinstance(n.func, ast.Name) and n.func.id=='PraxisInbound' for k in n.keywords if k.arg=='poll_timeout']; print(len(kw)==1 and isinstance(kw[0].value, ast.Constant) and kw[0].value.value==0.0)"` prints `True`.

F. Boundary-computation behaviour (deterministic value check):

- `python -c "from backtest_simulator.launcher.replay_clock import compute_kline_boundaries; from datetime import datetime, timezone; b=compute_kline_boundaries(window_start=datetime(2026,4,4,0,0,tzinfo=timezone.utc), window_end=datetime(2026,4,4,23,59,tzinfo=timezone.utc), interval_seconds=14400); print([t.isoformat() for t in b])"` prints `['2026-04-04T04:00:00+00:00', '2026-04-04T08:00:00+00:00', '2026-04-04T12:00:00+00:00', '2026-04-04T16:00:00+00:00', '2026-04-04T20:00:00+00:00']`.

G. Bookkeeping:

- `python -c "import json, pathlib; b=json.loads(pathlib.Path('.github/module_budgets.json').read_text()); print(isinstance(b.get('backtest_simulator/launcher/replay_clock.py'), int))"` prints `True`.
- `python -c "import re,pathlib; lines=[l for l in pathlib.Path('CHANGELOG.md').read_text().splitlines() if l.strip()][:1]; print(bool(re.match(r'#\s+v2\.5\.0\b', lines[0])))"` prints `True`.
- `python -c "import tomllib,pathlib; print(tomllib.loads(pathlib.Path('pyproject.toml').read_text())['project']['version'])"` prints `2.5.0`.

H. Behavioural assertions — each runs one named test from `tests/launcher/test_replay_clock.py` and pins the runtime contract; the test name itself is the assertion's regression-gate handle:

- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_calls_tick_once_per_boundary_per_sensor -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_advances_freezer_to_each_boundary_in_order -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_drains_submits_and_outcomes_to_quiescence_repeat_until_fixed_point -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_fails_loud_on_heterogeneous_cadence -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_fails_loud_when_predict_loop_running -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_fails_loud_when_outcome_loop_running -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_is_deterministic_across_runs -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestDriveWindow::test_drain_waits_for_delayed_routed_outcome_before_advancing -q --no-header --tb=no | tail -1` ends with ` 1 passed`.
- `uv run pytest tests/launcher/test_replay_clock.py::TestModuleImportClean::test_import_replay_clock_from_clean_interpreter -q --no-header --tb=no | tail -1` ends with ` 1 passed`.

---

## Current State

>
> Accepted:
>
> Rejected:

- `git show origin/main:backtest_simulator/launcher/clock.py | grep -c '_frozen_aware_timer_run'` returns at least `2`.
- `git show origin/main:backtest_simulator/launcher/launcher.py | grep -c 'def _advance_clock_until'` returns `1`.
- `git show origin/main:backtest_simulator/launcher/launcher.py | grep -c 'PraxisInbound(outcome_queue=nexus_outcome_queue)'` returns `1`.
- `git ls-tree origin/main -- backtest_simulator/launcher/replay_clock.py | wc -l | tr -d ' '` returns `0`.
- `python -c "import json,subprocess; t=subprocess.check_output(['git','show','origin/main:.github/module_budgets.json']); b=json.loads(t); print(b.get('backtest_simulator/launcher/replay_clock.py'))"` prints `None`.
- `git show origin/main:pyproject.toml | grep -E '^version = '` prints `version = "2.4.0"`.
- `git show origin/main:pyproject.toml | grep -c 'vaquum-nexus @ git+https://github.com/Vaquum/Nexus'` returns at least `1` (sibling lib is intentionally unpinned in `pyproject.toml`; every install pulls latest `main`; `uv.lock` is gitignored so it cannot be cited as repo state).
- `git show origin/main:tests/launcher/test_clock.py | grep -c '_frozen_aware_timer_run'` returns at least `1`.

---

## Target State

> predicate.
>
> Accepted:

- `git show HEAD:backtest_simulator/launcher/clock.py | grep -c '_frozen_aware_timer_run'` returns `0`.
- `git show HEAD:backtest_simulator/launcher/launcher.py | grep -c 'def _advance_clock_until'` returns `0`.
- `git ls-tree HEAD -- backtest_simulator/launcher/replay_clock.py | wc -l | tr -d ' '` returns `1`.
- `git show HEAD:backtest_simulator/launcher/launcher.py | grep -cE '^from backtest_simulator\\.launcher\\.replay_clock import'` returns at least `1`.
- `python -c "import json,subprocess; t=subprocess.check_output(['git','show','HEAD:.github/module_budgets.json']); b=json.loads(t); print(isinstance(b.get('backtest_simulator/launcher/replay_clock.py'), int))"` prints `True`.
- `git show HEAD:pyproject.toml | grep -E '^version = '` prints `version = "2.5.0"`.
- `git show HEAD:tests/launcher/test_clock.py | grep -c '_frozen_aware_timer_run'` returns `0`.
- `uv run pytest tests/launcher/test_replay_clock.py tests/launcher/test_clock.py -q` exits `0`.

---

## Surfaces

- `backtest_simulator/launcher/clock.py`
- `backtest_simulator/launcher/launcher.py`
- `pyproject.toml`
- `CHANGELOG.md`
- `tests/launcher/test_clock.py`
- `tests/tools/test_lint_ci_contract.py`

Added:
- `backtest_simulator/launcher/replay_clock.py`
- `tests/launcher/test_replay_clock.py`

Removed:
-

---

## Signature Diff

### `backtest_simulator/launcher/clock.py` :: module-level monkey-patch
Before:
```python
def _frozen_aware_timer_run(self: threading.Timer) -> None:
    ...
    while True:
        if self.finished.wait(_TIMER_POLL_INTERVAL_SECONDS):
            return
        now = datetime.now(UTC)
        if target is None or now < target - interval * 2:
            ...
        if now >= target:
            break
    if not self.finished.is_set():
        fn(*fn_args, **fn_kwargs)
    self.finished.set()


setattr(threading.Timer, 'run', _frozen_aware_timer_run)
```
After:
```python
# (function and setattr removed; threading.Timer.run is no longer
# monkey-patched.)
```

### `backtest_simulator/launcher/launcher.py` :: `BacktestLauncher._advance_clock_until`
Before:
```python
def _advance_clock_until(self, end: datetime, freezer: object) -> None:
    real_start = os.times()[4]
    kline_size_s = self._clock_tick_seconds.total_seconds()
    use_smart = kline_size_s > 120.0
    ...
    while datetime.now(UTC) < end:
        ...
        tick_fn = getattr(freezer, 'tick')
        if use_smart:
            ...
            tick_fn(timedelta(seconds=tick_amt))
            time.sleep(pause_s)
        else:
            tick_fn(self._clock_tick_seconds)
            time.sleep(_CLOCK_TICK_REAL_PAUSE_SECONDS)
        self._drain_pending_submits()
    ...
```
After:
```python
# Removed. BacktestLauncher.run_window now delegates window replay
# to ReplayClock.drive_window from
# backtest_simulator.launcher.replay_clock.
```

### `backtest_simulator/launcher/replay_clock.py` :: new module (functions and class)
Before:
```python
# (file does not exist on origin/main)
```
After:
```python
def compute_kline_boundaries(
    *,
    window_start: datetime,
    window_end: datetime,
    interval_seconds: int,
) -> list[datetime]:
    '''Return every epoch-aligned `interval_seconds` boundary in
    (window_start, window_end].

    The interval is half-open at the start: when `window_start` lies
    exactly on an epoch-aligned boundary, the first returned tick is
    `window_start + interval_seconds`. This matches the runtime
    contract that PredictLoop's first tick after start fires one
    interval after the boundary at which the loop began.
    '''


@dataclass(frozen=True)
class ReplayClock:
    '''Single-thread schedule-driven replay clock for one backtest window.

    Drives one (window_start, window_end] window through:
      1. boundaries = compute_kline_boundaries(...)
      2. for each boundary in boundaries:
           - freezer.move_to(boundary)
           - for each wired sensor: predict_loop.tick_once(wired)
           - drain to quiescence (REPEAT UNTIL FIXED POINT):
               while True:
                   drain_pending_submits()  # caller-provided
                   consumed = False
                   while outcome_loop.tick_once():
                       consumed = True
                   if not consumed:
                       break
             — `on_outcome` may emit new actions, which produce new
             submits, which in turn produce new outcomes; the loop
             only exits when both queues are empty in the same pass.

           Drain contract for the caller-provided
           `drain_pending_submits` callable: it MUST pump the asyncio
           event loop until every command delivered to the venue
           adapter has been routed through `_make_outcome_router` into
           the `nexus_outcome_queue` (i.e., delivered == routed). A
           simple "delivered count >= submitted count" check is
           insufficient because Praxis may have processed the submit
           before the router's awaitable resolved. The launcher's
           wrapper for this callable in this slice satisfies that
           contract by yielding to the running asyncio loop until the
           router's per-account queue length matches the per-account
           outbound delivered count.
      3. freezer.move_to(window_end); drain to quiescence (same loop)

    Fail-loud invariants enforced inside drive_window:
      - all wired sensors share one interval_seconds (raises
        ValueError on heterogeneous cadence).
      - predict_loop.running is False on entry (raises RuntimeError).
      - outcome_loop.running is False on entry (raises RuntimeError).
    '''

    def drive_window(
        self,
        *,
        window_start: datetime,
        window_end: datetime,
        wired_sensors: Sequence[WiredSensor],
        predict_loop: PredictLoop,
        outcome_loop: OutcomeLoop,
        drain_pending_submits: Callable[[], None],
        freezer: FrozenDateTimeFactory,
    ) -> None: ...
```

### `backtest_simulator/launcher/launcher.py` :: `PraxisInbound` construction site
Before:
```python
praxis_inbound = PraxisInbound(outcome_queue=nexus_outcome_queue)
```
After:
```python
praxis_inbound = PraxisInbound(
    outcome_queue=nexus_outcome_queue,
    poll_timeout=0.0,
)
```

### `backtest_simulator/launcher/launcher.py` :: lifecycle calls (removed)
Before:
```python
predict_loop.start()
...
outcome_loop.start()
...
outcome_loop.stop()
```
After:
```python
# All three calls removed from the BacktestLauncher path. ReplayClock
# drives both loops via tick_once; neither background worker is ever
# started in the simulator.
```

### `tests/launcher/test_clock.py` :: imports and tests of the monkey-patch
Before:
```python
from backtest_simulator.launcher.clock import (
    _frozen_aware_timer_run,
    accelerated_clock,
)


def test_threading_timer_is_permanently_patched() -> None:
    ...
    assert threading.Timer.run is _frozen_aware_timer_run


def test_timer_cancel_stops_wait() -> None:
    ...
```
After:
```python
from backtest_simulator.launcher.clock import accelerated_clock


# `test_threading_timer_is_permanently_patched` and
# `test_timer_cancel_stops_wait` removed: the monkey-patch they
# pinned no longer exists. The remaining accelerated_clock tests
# (pin, re-entry guard, exception-release) stay.
```

---

## Tests That Prove the MVC

Every MVC line in the *Minimum Viable Capability* section maps to exactly one test in `tests/launcher/test_replay_clock.py`. Sections A–G are structural pins (AST / file / JSON / version) and each is encoded as one `test_*` with the same content as the MVC predicate. Section H lines ARE pytest invocations of named tests; the test name in the table below is the same name the MVC line cites.

| MVC section / assertion (shortened) | test file | test name |
|---|---|---|
| A: `replay_clock.py` defines `compute_kline_boundaries` and class `ReplayClock` | `tests/launcher/test_replay_clock.py` | `TestModuleStructure::test_module_defines_compute_kline_boundaries_and_replay_clock` |
| A: `compute_kline_boundaries` is keyword-only `(window_start, window_end, interval_seconds)` | `tests/launcher/test_replay_clock.py` | `TestModuleStructure::test_compute_kline_boundaries_keyword_only_parameters` |
| A: `ReplayClock.drive_window` keyword-only parameters in canonical order | `tests/launcher/test_replay_clock.py` | `TestModuleStructure::test_drive_window_keyword_only_parameters_in_canonical_order` |
| B: `clock.py` no longer defines `_frozen_aware_timer_run` | `tests/launcher/test_replay_clock.py` | `TestSurfacesRemoved::test_clock_no_longer_defines_frozen_aware_timer_run` |
| B: `clock.py` no longer patches `threading.Timer` (setattr OR attribute-assign) | `tests/launcher/test_replay_clock.py` | `TestSurfacesRemoved::test_clock_no_longer_patches_threading_timer` |
| B: `launcher.py` no longer defines `_advance_clock_until` | `tests/launcher/test_replay_clock.py` | `TestSurfacesRemoved::test_launcher_no_longer_defines_advance_clock_until` |
| C: `launcher.py` imports the `replay_clock` module | `tests/launcher/test_replay_clock.py` | `TestLauncherUsesReplayClock::test_launcher_imports_replay_clock` |
| C: `BacktestLauncher.run_window` body calls `.drive_window(...)` with kwargs covering `{window_start, window_end, wired_sensors, predict_loop, outcome_loop, drain_pending_submits, freezer}` | `tests/launcher/test_replay_clock.py` | `TestLauncherUsesReplayClock::test_run_window_calls_drive_window_with_required_kwargs` |
| C: `launcher.py` does not call `predict_loop.start()` or `outcome_loop.start()` | `tests/launcher/test_replay_clock.py` | `TestLauncherUsesReplayClock::test_launcher_does_not_start_predict_or_outcome_loop` |
| D: `launcher.py` does not construct `TimerLoop(...)` | `tests/launcher/test_replay_clock.py` | `TestTimerLoopFailLoud::test_launcher_does_not_construct_timer_loop` |
| D: `launcher.py` raises `RuntimeError` mentioning `timer_specs` (fail-loud guard) | `tests/launcher/test_replay_clock.py` | `TestTimerLoopFailLoud::test_launcher_raises_runtime_error_on_non_empty_timer_specs` |
| E: `launcher.py` constructs `PraxisInbound(..., poll_timeout=0.0)` | `tests/launcher/test_replay_clock.py` | `TestLauncherUsesReplayClock::test_launcher_constructs_praxis_inbound_with_zero_poll_timeout` |
| F: `compute_kline_boundaries` returns 5 epoch-aligned 4h boundaries on 2026-04-04 | `tests/launcher/test_replay_clock.py` | `TestComputeKlineBoundaries::test_one_day_4h_returns_five_boundaries` |
| G: `module_budgets.json` includes a budget for `replay_clock.py` | `tests/launcher/test_replay_clock.py` | `TestBookkeeping::test_module_budgets_includes_replay_clock` |
| G: `CHANGELOG.md` top header is `v2.5.0` | `tests/launcher/test_replay_clock.py` | `TestBookkeeping::test_changelog_first_version_header_is_2_5_0` |
| G: `pyproject.toml` `[project].version` is `2.5.0` | `tests/launcher/test_replay_clock.py` | `TestBookkeeping::test_pyproject_version_is_2_5_0` |
| H: `drive_window` calls `predict_loop.tick_once(wired)` once per boundary per sensor | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_calls_tick_once_per_boundary_per_sensor` |
| H: `drive_window` advances freezer to each boundary in monotonic order | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_advances_freezer_to_each_boundary_in_order` |
| H: `drive_window` drains submits + outcomes to quiescence (repeat-until-fixed-point) | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_drains_submits_and_outcomes_to_quiescence_repeat_until_fixed_point` |
| H: `drive_window` raises `ValueError` on heterogeneous-cadence wired sensors | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_fails_loud_on_heterogeneous_cadence` |
| H: `drive_window` raises `RuntimeError` when entered with `predict_loop.running == True` | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_fails_loud_when_predict_loop_running` |
| H: `drive_window` raises `RuntimeError` when entered with `outcome_loop.running == True` | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_fails_loud_when_outcome_loop_running` |
| H: two `drive_window` calls with identical inputs produce identical recorded `tick_once` sequences | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_is_deterministic_across_runs` |
| H: drain-to-quiescence waits for an outcome that the router enqueues only AFTER the first submit-drain returns (delivered-but-not-routed barrier) | `tests/launcher/test_replay_clock.py` | `TestDriveWindow::test_drain_waits_for_delayed_routed_outcome_before_advancing` |
| H: `import backtest_simulator.launcher.replay_clock` succeeds in a clean subprocess | `tests/launcher/test_replay_clock.py` | `TestModuleImportClean::test_import_replay_clock_from_clean_interpreter` |

---

## Regression Gate

>
> Accepted:
>   on non-zero exit.
>
> Rejected:

Workflow file: `.github/workflows/pr_checks_lint.yml`
Job name: `pr_checks_lint` (the "Coverage floor gate" step)
Specific failing command the job runs (verbatim from the workflow):
```
.venv-lint/bin/python -m coverage run --branch --source=backtest_simulator -m pytest tests/ -q
.venv-lint/bin/python -m coverage json -o coverage.json
.venv-lint/bin/python tools/check_coverage_floor.py
```

How the gate catches regressions of THIS slice's MVC:
- Every MVC assertion in section H is encoded as a test in `tests/launcher/test_replay_clock.py`. A PR that regresses the underlying behaviour breaks the test, which fails `pytest tests/`.
- Every MVC assertion in sections A–G is structural (AST/file/JSON/version checks). The same test file pins each one as a separate `test_*` so it shows up in the same `pytest tests/` run.

Honest limitation (no overstated claim): if a future PR silently DELETES `tests/launcher/test_replay_clock.py` itself, this gate alone does not flag the deletion. The Surfaces and Out of Scope contracts (slice-scope CI gate `pr_checks_slice`) are what mechanically enforce that future PRs touching `tests/launcher/test_replay_clock.py` must declare it in their own slice's Surfaces, but a PR that removes the file without re-adding it to its Surfaces will be caught by the slice-scope check on that PR, not by this slice's regression gate. Stronger deletion-proof guards (e.g., adding a structural CI test under `tests/tools/`) are deferred to a follow-up slice rather than smuggled into this one (out of scope).

---

## Out of Scope

- `backtest_simulator/cli/**`
- `backtest_simulator/feed/**`
- `backtest_simulator/honesty/**`
- `backtest_simulator/launcher/__init__.py`
- `backtest_simulator/launcher/action_submitter.py`
- `backtest_simulator/launcher/poller.py`
- `backtest_simulator/sensors/**`
- `backtest_simulator/strategies/**`
- `backtest_simulator/venue/**`
- `backtest_simulator/pipeline/**`
- `backtest_simulator/exceptions.py`
- `backtest_simulator/__init__.py`
- `backtest_simulator/_limen_cache.py`
- `backtest_simulator/determinism.py`
- `backtest_simulator/wall_clock.py`
- `tests/cli/**`
- `tests/feed/**`
- `tests/honesty/**`
- `tests/integration/**`
- `tests/pipeline/**`
- `tests/sensors/**`
- `tests/strategies/**`
- `tests/tools/**` (except `tests/tools/test_lint_ci_contract.py` which pins the ruff per-file-ignores dict modified in this slice)
- `tests/venue/**`
- `tools/**`
- `.github/workflows/**`
- `.github/rulesets/**`

---

## Risks and Open Questions

Risks:
- The Nexus circular import at module-import time (`nexus.startup.shutdown_sequencer` <-> `nexus.strategy.predict_loop`) means `ReplayClock` must defer the `from nexus.strategy.predict_loop import PredictLoop` to inside the method body, not module top, otherwise importing `replay_clock.py` from a clean Python session raises `ImportError`. The slice contains a unit test that imports `replay_clock` from a clean interpreter and asserts no `ImportError` propagates.
- `BacktestLauncher._drain_pending_submits` is private. The replay-clock module accepts a `drain_pending_submits: Callable[[], None]` parameter rather than reaching into the launcher; the launcher passes `self._drain_pending_submits` at construction. No `_`-prefix reach-through outside the launcher itself.

Open questions:
- After this slice merges and the cleanup begins, should `clock.py` be folded into `replay_clock.py` entirely (eliminating `accelerated_clock` as a separate context manager)? Recommendation: defer to a follow-up slice once the cleanup confirms no other callers of `accelerated_clock`.

---

## Closeout

> bot verifies:

Merge SHA:
Merged PR number:
- pr_checks_slice :
- pr_checks_cc :
- pr_checks_typing :
- pr_checks_fail_loud :
- pr_checks_version :
- pr_checks_lint :
- pr_checks_ruleset :
- PR Checks CodeQL :

mikkokotila:
Closes #64. Implements [RFC-2001](https://github.com/Vaquum/backtest_simulator/issues/65).

## Summary

- Replace the racy `threading.Timer.run` monkey-patch in `backtest_simulator/launcher/clock.py` and `_advance_clock_until` in `backtest_simulator/launcher/launcher.py` with a single-thread, schedule-driven replay clock at `backtest_simulator/launcher/replay_clock.py`.
- `BacktestLauncher.run_window` now calls `ReplayClock().drive_window(...)`, which jumps frozen time to each epoch-aligned kline boundary, calls `predict_loop.tick_once(wired)` per sensor (Nexus 0.41.0 public API from [Vaquum/Nexus#61](https://github.com/Vaquum/Nexus/pull/61)), and drains submits + outcomes to quiescence (repeat-until-fixed-point) before advancing.
- Neither `PredictLoop.start()` nor `OutcomeLoop.start()` is called on the launch thread anymore; both background loops are driven synchronously via their public `tick_once` entries. `PraxisInbound(... poll_timeout=0.0)` so nonblocking outcome drains cost no real time per empty pass. The launcher fails loud if `sequencer.timer_specs` is non-empty (TimerLoop not supported in the synchronous replay path; follow-up RFC required).
- `_drain_pending_submits` now waits for `routed >= delivered` (a new routed-outcomes counter incremented inside `_make_outcome_router`) — closes the asyncio "delivered != routed" gap.

## Validation — `bts sweep` is now deterministic

The point of slice 0 is the canonical command running deterministically. Three back-to-back canonical runs (`slice0-run-A`, `slice0-run-B`, `clock-A`) produce **byte-equal** `sweep_per_window.csv` AND `sweep_per_tick.csv`:

```
diff slice0-run-A vs slice0-run-B per_window  → IDENTICAL
diff slice0-run-A vs clock-A per_window       → IDENTICAL
diff slice0-run-A vs clock-A per_tick         → IDENTICAL
```

Side effects of the race that are now gone:
- `ParityViolation` raised on every pre-slice-0 run on the same args. Post slice 0: `signals parity OK n_compared=75` on every run.
- Tick timestamps land on exact 4h boundaries (`04:00:00`, `08:00:00`, ..., `20:00:00`); the race-induced `:01` and `:59:00` artefacts are gone.
- Replay batch wall time ~26s (was ~30s+).

The canonical command exercised:

```
.venv/bin/bts sweep \
  --bundle ~/dev/_examples/BtcLogRegEVSFD__r0014.zip \
  --n-decoders 3 --n-permutations 5000 \
  --trading-hours-start 00:00 --trading-hours-end 23:59 \
  --replay-period-start 2026-04-01 --replay-period-end 2026-04-05 \
  --max-allocation-per-trade-pct 0.4 --predict-lookback 1 \
  --cpcv-n-groups 4 --cpcv-n-test-groups 2 --cpcv-purge-seconds 0 --cpcv-embargo-seconds 0 \
  --session-id <id>
```

## Surfaces

| File | Change |
|---|---|
| `backtest_simulator/launcher/replay_clock.py` | new — 250 significant LOC |
| `backtest_simulator/launcher/clock.py` | dropped `_frozen_aware_timer_run` + `setattr(threading.Timer, ...)`; kept `accelerated_clock` and `tick_frozen_time` |
| `backtest_simulator/launcher/launcher.py` | dropped `_advance_clock_until`, `predict_loop.start()`, `outcome_loop.start()`, conditional `TimerLoop` construction; added `ReplayClock` import + `run_window` delegation, `timer_specs` fail-loud raise, `PraxisInbound(poll_timeout=0.0)`, routed-outcomes counter |
| `tests/launcher/test_replay_clock.py` | new — 34 tests covering MVC sections A-H |
| `tests/launcher/test_clock.py` | dropped `test_threading_timer_is_permanently_patched` and `test_timer_cancel_stops_wait`; remaining 3 accelerated_clock tests stay |
| `.github/module_budgets.json` | added `replay_clock.py` entry (260); `clock.py` budget 180 → 100 |
| `pyproject.toml` | 2.4.0 → 2.5.0 |
| `CHANGELOG.md` | v2.5.0 entry |

## Test plan

- [x] `pytest tests/launcher/test_replay_clock.py tests/launcher/test_clock.py -q` — 34 new + 3 existing pass
- [x] Full launcher suite: `pytest tests/launcher/ -q` — 57/57 pass
- [x] `ruff check backtest_simulator/launcher/ tests/launcher/` — clean
- [x] `tools/check_module_budgets.py` — PASS
- [x] Three canonical `bts sweep` runs produce byte-equal outputs (validation above)
- [ ] CI green on this PR

---

[budget-raise: backtest_simulator/launcher/replay_clock.py: round-2 added drain-to-quiescence wall-clock cap (`real_time_cap_seconds` field + deadline plumbing + `ReplayDeadlineExceededError` exception class) per zero-bang's review on the unbounded `while True`; budget 260 -> 270]


backtest_simulator/launcher/launcher.py
```diff
@@ -926,6 +937,18 @@ def __init__(
         self._submit_lock = threading.Lock()
         self._venue_adapter = venue_adapter
         self._max_allocation_per_trade_pct = max_allocation_per_trade_pct
+        # Routed-outcome counter incremented by `_make_outcome_router` after
+        # a translated Nexus outcome lands in the per-account queue. Paired
+        # with `_delivered_command_count` in `_drain_pending_submits` so
+        # the synchronous replay drain waits for delivered == routed,
+        # closing the asyncio router's tail.
+        self._routed_outcomes = 0
```
bit-mis, line 929: P1 — `clock_tick_seconds` is now a silent no-op. The constructor still accepts it (line 884), still validates it (rejects 0/negative at line 915), and still stores `self._clock_tick_seconds` (line 925), but nothing in this file reads `_clock_tick_seconds` after `_advance_clock_until` was deleted (`grep _clock_tick_seconds launcher.py` returns only the storage site). `backtest_simulator/cli/_run_window.py:430` actively passes `clock_tick_seconds=interval_seconds` with a comment claiming it makes the loop "jump big between boundaries... 250ms real-time pause" — none of which is true anymore. Either drop the parameter (callers get a TypeError instead of a silent contract loss) or wire it into `ReplayClock`. The same incomplete cleanup leaves the module-level constants `_REAL_TIME_CAP_SECONDS` (line 78) and `_CLOCK_TICK_REAL_PAUSE_SECONDS` (line 104) defined-but-unused; both should go with the parameter.

backtest_simulator/launcher/launcher.py
```diff
@@ -1422,10 +1490,17 @@ def run_window(self, start: datetime, end: datetime) -> None:
         try:
             with accelerated_clock(start) as freezer:
                 try:
-                    self._advance_clock_until(end, freezer)
+                    pl, ol, ws = self._predict_loop, self._outcome_loop, self._wired_sensors
+                    if pl is None or ol is None or ws is None:
+                        raise RuntimeError('run_window: predict_loop / outcome_loop / wired_sensors were not stashed by _run_my_nexus_instance')
+                    ReplayClock().drive_window(
```
bit-mis, line 1484: P2 — the `run_window` docstring (lines 1432-1461) still describes the deleted `_advance_clock_until` flow: it claims step 4 is "Main thread blocks until `datetime.now(UTC) >= end`, then calls `request_stop()`", references `_advance_clock_until` ticking the freezer by `_CLOCK_TICK_SECONDS`, and points readers to a "frozen-aware patch in `clock.py`" that no longer exists. The actual contract after this PR is the headline change (ReplayClock pre-computes epoch-aligned boundaries, `freezer.move_to`s each one, calls `tick_once(wired)` per sensor, drains submits+outcomes to a fixed point, then a final `move_to(window_end)` + drain). The docstring on the entry point is the first place a reader looks; it has to match. While there, the comment in `_record_submitted_command` at line 1562 also still refers to `_advance_clock_until` for drain coordination — same fix.

bit-mis (changes requested):
Change requested in inline comments.

backtest_simulator/launcher/launcher.py
zero-bang: +1 — also note that the module-level constants `_CLOCK_TICK_SECONDS` (line 103) and the long comment block at lines 91-104 explaining the tick sizing rationale are dead for the same reason. All three (`_CLOCK_TICK_SECONDS`, `_REAL_TIME_CAP_SECONDS`, `_CLOCK_TICK_REAL_PAUSE_SECONDS`) plus the parameter, its validation, and `self._clock_tick_seconds` should go in one pass.

backtest_simulator/launcher/launcher.py
zero-bang: +1 — specifically, line 1447 says "PredictLoop.start() has been called" which is now actively untrue (the PR's central invariant is that `predict_loop.start()` is NOT called). The step 3-4 description ("\_advance\_clock\_until ticks the freezer by \_CLOCK\_TICK\_SECONDS per iteration while asyncio sleeps and threading.Timer.run wait…") should be replaced with the ReplayClock contract: pre-computed epoch-aligned boundaries, `freezer.move\_to` per boundary, `tick\_once(wired)` per sensor, drain to quiescence, final `move\_to(window\_end)` + drain.

backtest_simulator/launcher/replay_clock.py
```diff
@@ -0,0 +1,299 @@
...
+    if window_end < window_start:
+        msg = (
+            f'compute_kline_boundaries: window_end ({window_end}) must be '
+            f'>= window_start ({window_start})'
+        )
+        raise ValueError(msg)
+
+    epoch = datetime(1970, 1, 1, tzinfo=window_start.tzinfo)
+    elapsed = int((window_start - epoch).total_seconds())
+    next_boundary_secs = (elapsed // interval_seconds + 1) * interval_seconds
```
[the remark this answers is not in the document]
zero-bang: +1 — concrete path: if `window_end` is naive and `window_start` is aware, line 113 (`window_end < window_start`) raises `TypeError: can't compare offset-naive and offset-aware datetimes` before the documented `ValueError` ever fires. Minimal fix: add `if window_end.tzinfo is None: raise ValueError(...)` between lines 112 and 113, matching the existing `window_start` guard.

zero-bang (commented):
Solid architectural change — the single-thread schedule-driven replay is a clear improvement over the racy `threading.Timer` path, and the deterministic `bts sweep` validation is compelling. The `_drain_to_quiescence` repeat-until-fixed-point design correctly handles the `on_outcome → new submit → new outcome` chain.

I agree with bit-mis's P1 (dead `clock_tick_seconds` parameter + dead module constants) and P2 (stale `run_window` docstring + stale `_record_submitted_command` comment), and with Copilot's `window_end.tzinfo` validation gap. Replied to all three threads with supporting detail.

One new finding below on the loss of the wall-time safety cap. The rest of the change looks correct — the happens-before via `_nexus_running` Event between the launch thread stashing handles and `run_window` reading them is sound, and the `routed >= delivered` drain phase correctly closes the asyncio router tail for the synchronous path.

backtest_simulator/launcher/replay_clock.py
```diff
@@ -0,0 +1,299 @@
...
+        """Repeat-until-fixed-point drain.
+
+        Each pass: settle pending submits, then consume every
+        currently-routed outcome. If consuming an outcome triggered
+        `on_outcome` to emit new actions, those produce new submits
+        in the next pass; the loop only exits when both queues are
+        empty in the same pass.
+        """
+
+        while True:
```
zero-bang, line 343: The old `_advance_clock_until` had a `_REAL_TIME_CAP_SECONDS` (600s) wall-time guard that logged a warning and returned when the replay window consumed too much real time. `drive_window` and `_drain_to_quiescence` have no equivalent.

Concrete failure mode: if a strategy's `on_outcome` handler emits a new `action_submit` whose fill triggers another `on_outcome` → `action_submit` (cyclic chain), each individual `drain_pending_submits()` call succeeds quickly (fresh `_DRAIN_TIMEOUT_SECONDS` budget per call), so the outer `while True` here never breaks and the process hangs silently. The old code would at least log `backtest window exceeded 600s of real wall time` and bail.

Suggestion: thread `os.times()[4]` through `drive_window` and raise (or log + return) when total real wall time exceeds a cap. The existing `_REAL_TIME_CAP_SECONDS` constant is still defined in `launcher.py` — if you keep it, wire it into `drive_window`; if you drop it per bit-mis's P1, add a cap parameter to `drive_window` instead.

backtest_simulator/launcher/replay_clock.py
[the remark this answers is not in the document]
mikkokotila: Addressed in e0ca8e0 — `compute_kline_boundaries` now validates `window_end.tzinfo is not None` (raises `ValueError` not `TypeError` for naive end) and `window_start.tzinfo == window_end.tzinfo` (rejects tz-mismatched windows that would otherwise produce off-by-N-hour epoch alignment). Two new tests pin the new error paths: `test_naive_window_end_raises`, `test_mismatched_tzinfo_raises`.

backtest_simulator/launcher/launcher.py
```diff
@@ -1422,10 +1490,17 @@ def run_window(self, start: datetime, end: datetime) -> None:
...
+                        window_start=start, window_end=end, wired_sensors=ws,
+                        predict_loop=pl, outcome_loop=ol,
+                        drain_pending_submits=self._drain_pending_submits, freezer=freezer,
+                    )
```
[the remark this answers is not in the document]
mikkokotila: Addressed in e0ca8e0 — `run_window`'s docstring rewritten to describe the new synchronous replay path (boot under real time → `_nexus_running` event → `accelerated_clock` → `ReplayClock.drive_window` jumps frozen time per pre-computed kline boundary, calls `predict_loop.tick_once(wired)` per sensor synchronously, drains submits and routed outcomes to quiescence) instead of the deleted `_advance_clock_until` + `threading.Timer.run` patch flow.

backtest_simulator/launcher/launcher.py
mikkokotila: Addressed in e0ca8e0 — `run_window`'s docstring rewritten to describe the new synchronous replay path (boot under real time → `_nexus_running` event → `accelerated_clock` → `ReplayClock.drive_window` jumps frozen time per pre-computed kline boundary, calls `predict_loop.tick_once(wired)` per sensor synchronously, drains submits and routed outcomes to quiescence) instead of the deleted `_advance_clock_until` + `threading.Timer.run` patch flow.

backtest_simulator/launcher/launcher.py
mikkokotila: Addressed in e0ca8e0 — `clock_tick_seconds` is now explicitly marked accepted-but-unused with a comment block naming slice 0 (#64) as the reason and pointing at the follow-up slice that should remove it. The constructor now also logs an INFO line at construction time when a non-default value is passed, so the operator sees the param being ignored on every run rather than discovering it via grep. Sub-zero validation is retained so a caller passing a bogus value still fails fast.

Full removal of `clock_tick_seconds` (and the matching `cli/_run_window.py:430` call site) is blocked by this slice's `≤ 3 package files` scope budget; touching `_run_window.py` would tip from 3 → 4 package files modified. The follow-up slice that does the removal can be sized small (one constructor-arg drop, one call-site edit, version bump) and will be picked up immediately after this lands.

backtest_simulator/launcher/replay_clock.py
zero-bang: Still not addressed. `_drain_to_quiescence` is an unbounded `while True` — if a strategy's `on_outcome` handler emits a new action whose fill triggers another `on_outcome` (cyclic chain), each individual `drain_pending_submits()` call completes within its own `_DRAIN_TIMEOUT_SECONDS` budget, so the outer loop never breaks and the process hangs silently. The old `_advance_clock_until` had a 600s real-time cap that at least logged and bailed. Either thread an `os.times()[4]` wall-time cap through `drive_window` (the constant `_REAL_TIME_CAP_SECONDS` is still defined in launcher.py) or add a `max_real_seconds` parameter to `drive_window` and raise when the budget is exhausted.

zero-bang (changes requested):
Some items from my previous review still need to be addressed. See my replies on the relevant threads.

backtest_simulator/launcher/launcher.py
mikkokotila: Addressed in f6fe021 — `_CLOCK_TICK_SECONDS` (line 103) and `_CLOCK_TICK_REAL_PAUSE_SECONDS` (line 104) removed, the long comment block at lines 91-104 with the tick-sizing rationale removed, and the orphan `self._clock_tick_seconds` storage in `__init__` dropped. `_REAL_TIME_CAP_SECONDS` kept and re-purposed as the default wall-clock cap on `ReplayClock.drive_window` (per the related drain-cap thread). Stale `_advance_clock_until` reference in `_record_submitted_command`'s docstring fixed too.

backtest_simulator/launcher/launcher.py
mikkokotila: Re-checked the `run_window` docstring on the current commit (`e0ca8e0`/`f6fe021`). The "PredictLoop.start() has been called" line and the `_advance_clock_until` / `_CLOCK_TICK_SECONDS` references were already replaced in `e0ca8e0`. The current docstring (lines 1431-1456) describes the synchronous flow: boot under real time → `_nexus_running` event → `accelerated_clock` → `ReplayClock.drive_window` jumps frozen time per kline boundary, calls `predict_loop.tick_once(wired)` per sensor synchronously, drains submits + routed outcomes to quiescence. Step 1 explicitly states `(NOT started — see _run_my_nexus_instance)`. No further edit needed.

backtest_simulator/launcher/replay_clock.py
mikkokotila: Addressed in f6fe021 — added a wall-clock deadline to the drain-to-quiescence loop. `ReplayClock` now takes a `real_time_cap_seconds: float = 600.0` field; `drive_window` computes `deadline = os.times()[4] + self.real_time_cap_seconds` once at entry and threads it into `_drain_to_quiescence`. Each pass of the inner `while True` checks `os.times()[4] > deadline` and raises a new `ReplayDeadlineExceededError` (subclass of `RuntimeError`) rather than spinning further.

Concrete failure mode pinned in test `test_drain_aborts_loud_on_cyclic_on_outcome_chain`: the test's drain mock enqueues a fresh outcome on every call, so the repeat-until-fixed-point loop never naturally exits. With `real_time_cap_seconds=0.5`, `drive_window` raises `ReplayDeadlineExceededError` matching `'cyclic on_outcome'`. Tightened threshold (60s, 600s, etc.) is configurable per call.

Default 600s matches the legacy `_advance_clock_until` cap so production sweeps see the same wall-clock budget. Constructor-injectable so tests can use a tiny cap without wall-clock waiting.

mikkokotila: Addressed in f6fe021 — added a wall-clock deadline to the drain-to-quiescence loop. `ReplayClock` now takes a `real_time_cap_seconds: float = 600.0` field; `drive_window` computes `deadline = os.times()[4] + self.real_time_cap_seconds` once at entry and threads it into `_drain_to_quiescence`. Each pass of the inner `while True` checks `os.times()[4] > deadline` and raises a new `ReplayDeadlineExceededError` (subclass of `RuntimeError`) rather than spinning further.

Concrete failure mode pinned in test `test_drain_aborts_loud_on_cyclic_on_outcome_chain`: the test's drain mock enqueues a fresh outcome on every call, so the repeat-until-fixed-point loop never naturally exits. With `real_time_cap_seconds=0.5`, `drive_window` raises `ReplayDeadlineExceededError` matching `'cyclic on_outcome'`. Tightened threshold (60s, 600s, etc.) is configurable per call.

Default 600s matches the legacy `_advance_clock_until` cap so production sweeps see the same wall-clock budget. Constructor-injectable so tests can use a tiny cap without wall-clock waiting.

backtest_simulator/launcher/replay_clock.py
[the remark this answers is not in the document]
mikkokotila: Already addressed in e0ca8e0 (your comment was filed before that fix landed): `compute_kline_boundaries` now validates `window_end.tzinfo is not None` AND `window_start.tzinfo == window_end.tzinfo`. Two new tests pin both error paths.

zero-bang (approved):

zero-bang (approved):

mikkokotila:
@bit-mis — your review on `0a39248` is the only thing now blocking merge (`reviewDecision=CHANGES_REQUESTED`). All inline threads are resolved and zero-bang has approved the latest commit (`411a28e`); could you re-review and either dismiss or approve?

Your two findings, with the addressing commits:

**P1 — `clock_tick_seconds` silent no-op** (resolved thread `PRRT_kwDOSJhSRs5__xUO`):

- `e0ca8e0` — explicit comment block marking the param accepted-but-unused, INFO log when a non-default value is passed, sub-zero validation kept.
- `f6fe021` — *fully* removed the orphan `self._clock_tick_seconds` storage and the dead `_CLOCK_TICK_SECONDS` / `_CLOCK_TICK_REAL_PAUSE_SECONDS` constants per zero-bang's +1.
- The constructor still accepts `clock_tick_seconds` for backward compat with `cli/_run_window.py:430`'s call site (out of scope here per slice 0's `≤ 3 package files` budget); follow-up slice will do the symmetric removal.

**P2 — `run_window` docstring stale** (resolved thread `PRRT_kwDOSJhSRs5__xUT`):

- `e0ca8e0` — docstring rewritten end-to-end. Step 1 now reads "PredictLoop / OutcomeLoop are constructed (NOT started — see `_run_my_nexus_instance`)"; step 3 says "ReplayClock.drive_window jumps frozen time per pre-computed kline boundary, calls `predict_loop.tick_once(wired)` per sensor synchronously"; step 4 says "After drive_window returns, `request_stop()` releases the launch thread. There is no `_advance_clock_until`, no `threading.Timer.run` monkey-patch". No remaining stale references; line 1447 specifically does not say "PredictLoop.start() has been called".

CI on `411a28e`: all 10 checks green (CodeQL, CodeQL python, cc, fail_loud, honesty, lint, ruleset, slice, typing, version). Canonical `bts sweep` byte-equal across three runs (`slice0-run-A`, `slice0-run-B`, `clock-A`).

bit-mis (approved):

bit-mis (approved):

